Issue #8, opened 2020-11-10
Packaging issues


StanczakDominik:
I like to keep things up to date on issues (I forget stuff a lot), so I'll add what we discussed today (and some other minor issues I forgot to raise) as todo items:

- [x] Make pip install build an importable package just the way `makefile aurora` does (trial/error optimization and comparison of the two build processes, looks okay)
- [x] make sure pip install pulls the package dependencies (it didn't for me)
- [x] add basic CI for the build step
- [ ] add basic tests through based on `examples/basic.py`

---

Hey, great talk yesterday! :)

I'm 90% sure I'd be able to automate the installation process into a single `conda install -c conda-forge aurora` (assuming the name is not taken). The reason why it'd be conda forge is because that's where community built packages generally live, and the main Anaconda repo (the big default) does not really accept contributions.

Would you like me to try? :)

StanczakDominik:
And I guess from there we could start working on a PyPI installation.

fsciortino:
Hey Dominik! Thank you very much :)

You might be right about that, but it seems to me that one should be able to upload to Anaconda Cloud [see this](https://docs.anaconda.com/anaconda-cloud/). I actually pushed a [previous package](https://anaconda.org/sciortino/pystrahl), but there's some issue that I'm running into with Aurora, which has a slightly more complicated directory structure. (NB: the "aurora" name is indeed already taken by another user, sadly, that's why I'm opting for "aurorafusion" -- but the package should still import as "aurora"). 

Most of the conda and PyPI setup for `aurorafusion` should already be done in this Github repo, but I'm running into issues during installations, apparently because the package .tar files are not unwrapped correctly. You can have a look at my most recent attempt here [on Anaconda](https://anaconda.org/sciortino/aurorafusion) or [on PyPI](https://pypi.org/project/aurorafusion/). If you try to download the package .tar file, you can see what it's supposed to look like, but when you try to install... for some reason it's not maintaining the package structure and that doesn't let the code be functional. 

If you have time, I'd love to get your thoughts on what might be going wrong. Would you be free during the conference lunch break today? If so, I can send you a Zoom link. 

Thanks for your offer to help! Greatly appreciated :)

StanczakDominik:
Hey, is tomorrow okay with you? I present on PlasmaPy today after the lunch break (which, in the EU timezones, should be considered a walk-your-dog-before-bed break :smile:), and I'm probably going to be busy fretting about that :)

fsciortino:
[...]

PR #9, opened 2020-11-17
Fix version file, package import on wheel-based installation (e.g. with `pip install .`)

StanczakDominik:
This solves two problems from https://github.com/fsciortino/Aurora/issues/8 :

1. the version file is now attached to the source package via
   package_data. This is achieved through a rather ugly hack - I just
   `recursive-include`d the whole `aurora` directory. I couldn't come to
   terms with numpy's distutils.setup, which didn't want to let me
   include single files into wheels (but `sdist` works fine, which is
   confusing), so I sidestepped the issue. The directory is small, so
   the size (~300kb) should not be a problem - but I'd welcome a better
   solution!
    * I also changed setup.py to refer to the version in the source
      directory;
    * and `__init__.py` now finds it robustly, no matter the location,
      through pkgutil.
    * I was then able to remove `.version` from the main package
      directory, to decrease the confusion.
    * I also took the liberty to rename it to `version` (no starting
      dot), so Linux doesn't treat it as a hidden file.
2. I was getting a "no extension named `aurora._aurora`" error on
   importing the package; so I changed the extension's name from
   `_aurora` to `aurora._aurora`. It didn't break anything I was able to
   test; but it stopped `_aurora` from popping up as an independent
   package, which is a boon in my eyes.


fsciortino (approved):
This looks great! Thank you very much  :)

fsciortino:
@StanczakDominik it seems that the current version works well with `pip install -e .`, but I'm still having issues installing from PyPI. The `version` file is not copied into the installation and, as a result, the code crashes with I try to `import aurora`...

Also, I pushed to master a clearer listing of requirements, but this isn't working quite right because packages like `omfit_eqdsk` and `omfit_gapy` are not on Anaconda, but they are available via PyPI. Conda does not seem to figure that out by itself.... so one needs to install those before installing `aurorafusion`, otherwise the `aurorafusion` installation will crash...

@StanczakDominik if you have any further thoughts on how to correct these issues, they would be very welcome!

StanczakDominik:
I take it we're talking about https://test.pypi.org/project/aurorafusion/1.0.11/ ?  It is weird that the `version` file doesn't appear there. If I understand correctly, you just did `python setup.py sdist` and uploaded the .tar.gz that gets built, right? Admittedly I just focused on the wheel build, which might have been risky.

I'll tinker with it some more tomorrow and we'll get to the bottom of this :)

fsciortino:
Yes, I've given up on uploading wheels (i.e. using `bdist_wheel`) because Aurora isn't a pure-python project and I was getting a headache from a million forums that give different indications on how to compile and upload wheels. So, using `python setup.py sdist` will do. 

The source distribution (.tar.gz file) seems to be built (and uploaded) correctly, but something doesn't work when I try to `pip install` from PyPI... I've tried to make sure that non-python files are added to any installation by modifying the `setup.py`, but with little luck. 

Let me know if you can make any progress! Thank you!

Issue #10, opened 2020-11-24
trouble installing aurora


jmcclena:
I get the following error trying to import aurora from python:

```mcclenaghan@F-C02Z321ML ~ % python                      
Python 3.7.9 (default, Sep  6 2020, 13:20:25) 
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import aurora
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/aurorafusion-1.0.11-py3.7-macosx-10.15-x86_64.egg/aurora/__init__.py", line 5, in <module>
    __version__ = pkgutil.get_data("aurora", "version").decode()
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/pkgutil.py", line 637, in get_data
    return loader.get_data(resource_name)
  File "<frozen importlib._bootstrap_external>", line 916, in get_data
FileNotFoundError: [Errno 2] No such file or directory: '/opt/local/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/aurorafusion-1.0.11-py3.7-macosx-10.15-x86_64.egg/aurora/version'
>>> 
```

My attempt to install aurora is:

```mcclenaghan@F-C02Z321ML Aurora % sudo python setup.py build --fcompiler=gfortran
Password:
running build
running config_cc
unifing config_cc, config, build_clib, build_ext, build commands --compiler options
running config_fc
unifing config_fc, config, build_clib, build_ext, build commands --fcompiler options
running build_src
build_src
building extension "aurora._aurora" sources
f2py options: []
  adding 'build/src.macosx-10.15-x86_64-3.7/build/src.macosx-10.15-x86_64-3.7/aurora/fortranobject.c' to sources.
  adding 'build/src.macosx-10.15-x86_64-3.7/build/src.macosx-10.15-x86_64-3.7/aurora' to include_dirs.
build_src: building npy-pkg config files
running build_py
running build_ext
customize UnixCCompiler
customize UnixCCompiler using build_ext
customize Gnu95FCompiler
Found executable /usr/local/bin/gfortran
customize Gnu95FCompiler using build_ext


mcclenaghan@F-C02Z321ML Aurora % sudo python setup.py install                   
running install
running bdist_egg
running egg_info
[...]

PR #11, opened 2020-11-24
Add compile/import tests!

StanczakDominik:
This tests whether the package compiles and imports successfully on every commit and pull request. It's failing right now for not-in-place installs, as shown by #10 & #9. Addresses #8 a bit.

I also added Numpy to setup_requires due to the f2py compilation step
requiring it.

This is not called with `python setup.py install/develop`, and has to be
done manually, so I added a step that install requirements beforehand.

I also added `pexpect` for omfit_commonclasses, which seemed to fail on
import for lack of it; this should be removed when possible and
double-checked with the great folks at OMFIT - seems like a missing
requirement :)

Fail-fast in the `strategy` part allows all the builds to proceed
independently. See https://github.community/t/build-got-was-canceled-unexpectedly/16130

I'll keep building on top of this and trying to fix the version file failing.


StanczakDominik:
I see Github won't display these tests until they're merged into master; if you'd like to take a look at how it would work, see https://github.com/StanczakDominik/Aurora/pull/1 .

requirements.txt
```diff
@@ -4,3 +4,4 @@ matplotlib
 xarray
 omfit_eqdsk
 omfit_gapy
+pexpect
```
fsciortino, line 7: @StanczakDominik I think that pexpect is needed by IPython, which isn't strictly necessary in Aurora, but is the way I personally tested the package `examples`. Thanks for adding it to the requirements!

StanczakDominik:
Gotcha!

So it turns out package_data works differently in numpy's distutils than I'm used to from setuptools. I should have started out by reading its documentation. This should now work nicely; there's a single place to set your version, and the Configuration object from numpy.distutils handles the version check "by default", to quote from https://numpy.org/doc/stable/reference/distutils.html:

> This method scans files named __version__.py, <packagename>_version.py, version.py, and __svn_version__.py for string variables version, __version__, and <packagename>_version, until a version number is found.

Which is cool and solves our problem!

I suspect we'll need *the power of distutils docs* again to grab package data, but that'll come once we adapt basic.py into a test, and I think this PR is now in a good place to review and merge, before it grows overlarge.

aurora/__init__.py
```diff
@@ -1,8 +1,7 @@
 """:py:mod:`aurora` 
 """
 
-import pkgutil
-__version__ = pkgutil.get_data("aurora", "version").decode()
+from .version import version as __version__
```
fsciortino, line 4: @StanczakDominik this is awesome!

setup.py
```diff
@@ -27,19 +28,11 @@
...
-
 aurora_dir = os.path.dirname(os.path.abspath(__file__))
-with open(os.path.join(aurora_dir, 'aurora', 'version')) as vfile:
-    version = vfile.read().strip()
-
 install_requires = open('requirements.txt').read().split('\n')
     
-setup(name=package_name,
-      version=version,
+config = Configuration(
```
fsciortino, line 34: @StanczakDominik is this needed because this Configuration class effectively finds the version file automatically, while the `setup` method doesn't?

StanczakDominik: You know, I haven't checked it with just setup and removing the version argument. Sounds like it should work! I'll check it out after some sleep :)

StanczakDominik:
Well, that worked (https://github.com/StanczakDominik/Aurora/pull/1). Whaddya know. :sweat_smile:

PR #13, opened 2020-11-25
Conda and PYPI

fsciortino:
Github workflow to automate/simplify pushing to Anaconda

@StanczakDominik thoughts?

### DIFF
--- a/.github/workflows/python-package-conda.yml
+++ b/.github/workflows/python-package-conda.yml
@@ -1,0 +1,31 @@
+name: Python Package using Conda
+
+on: [push]
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 5
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.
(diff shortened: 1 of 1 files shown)


StanczakDominik:
You mentioned earlier that you push Aurora to the main Anaconda repository, rather than conda-forge as we do. I don't know much about the former, but the latter has a pretty amazing set up: when they detect a new version on PyPI, they mostly automatically submit a PR to your "feedstock" (think conda build recipe) to upgrade the package there. 

So if we move there and tackle #14 first, this comes for free, basically. But it'd take the transition first.

What do you think?

StanczakDominik:
https://github.com/conda-forge/plasmapy-feedstock is the feedstock for PlasmaPy if you'd like to take a look.

fsciortino:
Pretty cool! Thanks for suggesting that. I am a little worried about adding complexity on complexity, because at some point it can all explode and become a time-eater, but I get the feeling that these are standard tools in the software-developing world that it's good to learn -- so I appreciate you bringing this up 👍 

So, let's do as you suggested, working on #14 and setting up the feedstock!

PR #14, opened 2020-11-25
PyPI workflow

fsciortino:
Similar to #13, but for PyPI. 

@StanczakDominik


StanczakDominik:
Fun fact, I was going to do this next up for PlasmaPy. For context: we already have infrastructure for azure pipelines to package stuff up when a tagged commit is pushed, but transitioning that to github actions should make things a little simpler. So it'd be easy to set up, and I can do both in parallel.

fsciortino:
@StanczakDominik could you check if you can see a "Github action secret" for this repo? I added my PyPI username, but not the password, just to make sure that I'm using this correctly.

StanczakDominik:
I think I'd first need to be added as a collaborator over at https://github.com/fsciortino/Aurora/settings/access - I get a 404 on trying to access the repositotry's settings.

fsciortino:
@StanczakDominik sorry, I thought I had done that already.

StanczakDominik:
Wait, lmao, this and #13 are pull requests. How the hell did I miss that :rofl:

StanczakDominik (approved):
Looks good! I thought I was going to do it due to the above mix-up, and I was going to use https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ . I like how that version also uploads to TestPyPI; then you could do that on push to master as well and be sure your released version is fine. But this looks good!

fsciortino:
I have little clue if this would work -- I basically copy/pasted it from the Github suggestion. Please let me know if you think that any changes would be needed. I've added a PyPI username and password as "Secrets" of the repo, so this `python-publish.yml` should be able to see those. 

I've also added some tests to the `meta.yaml` file, which should be run when uploading to conda-forge. My understanding is that this complements the `tests.yml` that you added, since it tests whether the package builds well on conda-forge, while your file tests (via Github, not conda-forge) whether the code works well at all.

fsciortino:
@StanczakDominik I agree that making use of TestPyPI would be nice to make sure that any update to master would work. Could you add that? 
It's not clear to me when the current version that I pushed here would get triggered...

StanczakDominik:
Based on

```yaml
on:
  release:
    types: [created]
```
I guess it'll act on creating a release over at https://github.com/fsciortino/Aurora/releases/new . I haven't really used that, though. The setup we have in PlasmaPy (lovingly borrowed straight from SunPy) uses pushing a `git tagged` commit to any branch on the main repo as its trigger.

StanczakDominik:
Looks like this is - roughly - going to work. Two issues to consider:

* the testPyPI upload is failing, most likely because the PyPI action is set up to take an auth token rather than your password, which you can set up at https://test.pypi.org/manage/account/ for testPyPI and have to replace the secret in the repository's settings.
* I didn't yet manage to get the actual upload to go through (no aurorafusion access on PyPI - I'll run a similar test on the PlasmaPy side to experiment). What I'm wondering about is - how is this going to discern between two commits with the same version (1.0.11 atm)?

StanczakDominik:
I've got some better ideas after doing a little experimenting, please don't merge it yet!

fsciortino:
OK. I was looking at PyPI tokens as you suggested -- I guess that those are still useful, so I'll set one up while you work on the new ideas!

fsciortino:
I'm changing the package version number because the current error that we're getting from Test PyPI reads
```
Checking dist/aurorafusion-0.0.0.tar.gz: PASSED, with warnings
  warning: `long_description_content_type` missing. defaulting to `text/x-rst`.
Uploading distributions to https://test.pypi.org/legacy/
Uploading aurorafusion-0.0.0.tar.gz

  0%|          | 0.00/139k [00:00<?, ?B/s]
100%|██████████| 139k/139k [00:00<00:00, 243kB/s]
NOTE: Try --verbose to see response content.
HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
This filename has already been used, use a different version. See https://test.pypi.org/help/#file-name-reuse for more information.
```
I'm currently not sure why it's trying to build aurorafusion-0.0.0.tar.gz -- the current version number isn't 0.0.0. Working on it...

fsciortino:
@StanczakDominik there is a problem with the automatic reading of the version number in the setup.py -- it seems that your procedure to let the version number be automatically read in isn't working at the moment. However, when I tried specifying the version number explicitely in setup.py (effectively duplicating the version number, present in two different files), things worked OK. 

I tried to use a setup.cfg to set the version number, but the tests fail in that case because they can't read the aurora.__version__ attribute since the Fortran part of the code has not bee compiled yet, so `_aurora` cannot be imported. This is a similar issue to what I encountered when putting documentation on ReadTheDocs -- I placed an if-statement in several Python files to prevent imports of `_aurora` when the word `sphinx` was appearing in `sys.modules`. I think that in this case, the issue with the version number could be solved by checking for something like
```
np.array(['something_that_is_only_present_when_setuptools_is_running' in k for k in sys.modules.keys()]).any()
```
but I can't quite figure out what is that "something" that should be checked for. I tried with "distutils", but that it didn't work and I was getting rather cryptic errors related to failed imports of basic packages like scipy...

It's pretty difficult to debug this stuff, but ... I think that we've made some good progress. It'd be great if you could find something better than this for version setting, but it's not critical. However, right now I think that the github workflow is trying to push a new version of the package to Test-PyPI and PyPI every time there is a push.... which seems a bit excessive, especially since this isn't even the master branch. I've never used "tags" and "releases", but it sounds like that's what we should use.

fsciortino:
Current status: setting the version in `setup.py` using `setup.cfg` is still failing. It doesn't seem worth spending more time on this at the moment -- it's OK to just repeat the same version number twice, in `aurora/version.py` and in `setup.py` itself. In principle, I would have expected to be able to just comment out the explicit version specification in `setup.py` at this point, but some cryptic errors are appearing when I do that. 

The more important issue is that when running basic testing for Aurora the `omfit_commonclasses` package is throwing an error due to `warnings` not being correctly imported in `gapy.py`. This has been corrected in OMFIT/unstable and just needs to be uploaded to PyPI @orso82 @smithsp.

fsciortino:
Quite a few things have gone into this branch and I think that it's better to merge with master at this stage. @StanczakDominik whenever you can make updates/upgrades to the testing/uploading infrastructure via another PR, those will be very much appreciated!

PR #15, opened 2020-11-25
Additional methods to specificy sources

fsciortino:
Features requested by @jmcclena to specify an arbitrary space- and time-dependent impurity neutral source.


fsciortino:
thank you very much @jmcclena for contributing to the iterative example for pellet modeling :)

PR #21, opened 2021-02-10
Updated solps module to correctly read fort.44 and fort.46 files

Reksoatr:


fsciortino (approved):

fsciortino (approved):
added reading of NATM, NMOL and NION in later push, useful to read neutral densities 
@Reksoatr

PR #22, opened 2021-02-13
switched to new omfit_classes package and modified user agreement

fsciortino:
Switch Aurora to use new `omfit_classes` package and specify that users cite OMFIT when using Python tools in Aurora. 

Outstanding issue before this is merged into master: `omfit_classes` currently prints to screen all the PYTHONPATH during the import. This is pretty annoying and can probably be avoided. @orso82 could you please have a look?

Also, did I get the OMFIT reference right in the Aurora user agreement? I took it from [the OMFIT user agreement](https://docs.google.com/forms/d/e/1FAIpQLSdHRDmTh5Kc0iVFKfUFonmC25C7DyMUIhlVFX2hh5XWyR2q4A/viewform).

### DIFF
--- a/USER_AGREEMENT.txt
+++ b/USER_AGREEMENT.txt
@@ -1,22 +1,22 @@
-Aurora User Agreement (2020.10.31) 
+Aurora User Agreement (2021.02.13) 
 
 By cloning, downloading or otherwise using Aurora sources, the recipient
 agrees:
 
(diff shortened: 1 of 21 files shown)


fsciortino:
thanks @orso82 for the update to `omfit_classes`. Everything seems to be fine now, so ready to merge.

PR #24, opened 2021-04-28
Imprad

odstrcilt:
validation of the code using pySTRAHL


fsciortino (approved):
Lots of good changes. There were a few issues with the nomenclature of variables for the ion source rates and the impurity ion name, but those were easily resolved. I've adapted the documentation as needed.

PR #27, opened 2021-05-10
correct use of interpolated source functions

fsciortino:
Simplified methods to interpolate time histories of particle sources. This is helpful to correct an error in ImpRad that occurs when trying to update source time histories using the source_utils.get_source_time_history function.


odstrcilt (commented):
Have you tested it? run_aurora_steady looks fine. You can let it run for 100s with increasing timestep. It is helpful if the transport coefficients are low and impurity is highly recycling.

PR #37, opened 2021-07-08
WIP: improve source handling in forward model

fsciortino:
This PR collects works to improve the way sources are handled in the Aurora forward model of particle transport. 

Main objectives:
- resolve awkwardness of sources that are explicitly given as a 2D matrix in (space,time) -- these should be handled in the same way as sources for which radial and time dependencies of sources are separable. 
- make puffing into the divertor an independent variable -- no reason to follow the STRAHL input convention of divertor particle sources being a fraction of core sources.
- allow specification of non-trivial recycling radial profile. This is an improvement wrt the assumption of recycling profiles being the same as the source radial profiles (which may also vary in time), since the latter could also represent a core impurity source (e.g. from a pellet).
- allow clearer accounting of particle numbers, including neutral particles rather than just ionized particles. 

This is work in progress (WIP).


aurora/source_utils.py
```diff
@@ -266,7 +266,13 @@ def get_radial_source(namelist, rvol_grid, pro_grid, S_rates,nt, Ti_eV=None):
         Radial profile of the impurity neutral source for each time step.
     '''
     r_src = namelist['rvol_lcfs'] + namelist['source_cm_out_lcfs']
-
+    nt = S_rates.shape[1]
+    try:
+        # TODO: invert order of dimensions of Ti_eV...
+        assert S_rates.shape==Ti_eV.T.shape
```
odstrcilt, line 272: When I have tried to run it from IMPRAD, it has crashed exactly here. You call this function from core.py as 

```
 self.rcl_rad_prof = source_utils.get_radial_source(
                    nml_rcl_prof, # namelist specifically to obtain exp decay from wall
                    self.rvol_grid, self.pro_grid,
                    self.S_rates[:,0,:],   # 0th charge state (neutral)
                    self._Ti[[0],:])
```

why are you using in argument self._Ti[[0],:] - this is a first timeslice of Ti. 
_get_radial_source_ should be working for time-evolving Ti profiles.

aurora/main.f90
```diff
@@ -295,27 +306,30 @@ subroutine run(  &
      if (sel_alg_opt.eq.0) then
         ! Use old algorithm, just for benchmarking
         call impden0( nion, ir, ra, rn,  &   !OUT: rn
-             diff, conv, par_loss_rates(:,it), src_rad_prof(:,it), &
+             diff, conv, par_loss_rates(:,it), source_2d(:,it), &
```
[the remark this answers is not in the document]
fsciortino: Good catch 👍

aurora/source_utils.py
fsciortino: I've corrected this now. 
I've made the recycling radial profile completely time INDEPENDENT because otherwise one would need to pass yet another large array to fortran/julia. This seems particularly inconvenient because most often users (a) may not even have recycling turned on, and (b) even when they do, they may want to set their own recycling profile (now possible with this new branch). It's just a choice led by pragmatism and simplicity. Let me know if you disagree.

aurora/main.f90
```diff
@@ -51,8 +49,12 @@ subroutine run(  &
   !                    This must be given for each charge state and time.
   !     par_loss_rates  real*8 (ir,nt)
   !                    Frequency for parallel loss on radial and time grids [1/s]
-  !     src_rad_prof real*8 (ir,nt)
-  !                    Radial profile of neutrals over time: n0(ir,t) = flx*src_rad_prof(ir,t).
+  !     source_2d real*8 (ir,nt)
+  !                    Radial profile of neutrals over time.
```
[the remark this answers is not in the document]
fsciortino: Have a look at how `source_2d` (or its individual time slices, called `src_prof` in impden.f90) is used here:
https://github.com/fsciortino/Aurora/blob/6fc8f3c82ebe39a1c030d3177916e42cdad45d54/aurora/impden.f90#L321
and here:
https://github.com/fsciortino/Aurora/blob/6fc8f3c82ebe39a1c030d3177916e42cdad45d54/aurora/impden.f90#L328

From these lines you see that the source radial profile is used to set the first array index of the impurity densities, i.e. the neutral stage. This is quite counter-intuitive -- maybe even wrong, one may say -- because sometimes we impose these source radial profiles by describing them as an ionization profile of neutrals.... but in fact, they are taken to represent the neutral stage themselves, rather than the first ionization stage. 

Looking at the lines above, I realized that the physical units of `source_2d` should be `particles`, not `particles/cm/s` or `particles/cm`. This is because each element of this array corresponds to the number of particles at each location and at each time. Would you agree??

aurora/main.f90
```diff
@@ -51,8 +49,12 @@ subroutine run(  &
...
+  !     rcl_rad_prof real*8 (ir)
+  !                    Radial distribution of impurities re-entering the core reservoir after recycling.
+  !                    NB: this should be a normalized profile! This condition is NOT enforced internally.
+  !                    If not provided, this is set to be equal to a normalized src_prof (the external source distribution)
```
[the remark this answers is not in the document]
fsciortino: `src_prof` is NOT normalized, it needs to give an absolute number of particles. Is the comment in main.f90 here not clear?

aurora/source_utils.py
odstrcilt: I need to check if it is not causing some issues with particle conservation. If not, than it is fine.

tomodstrcil:
Now I can run the code, but when I do it, in a few timesteps the density diverges and it is infinite. It is working for you? I have use the default setting in IMPRAD, default analytical profiles, default source (constant in time).

odstrcilt:
@fsciortino  still, there is some bug in recycling when I set R = 0. Any value of R > 0 works. Do you know where could be this bug?

odstrcilt:
Another issue, if I need to vary a source during the inference, I will have to call the function 'setup_kin_profs_depts'  or at least 'get_radial_source' in every iteration.

fsciortino:
@odstrcilt I see the issue. I'm working on it.

aurora/core.py
```diff
@@ -265,7 +268,7 @@ def setup_kin_profs_depts(self):
         self.total_source = np.pi*np.sum(self.src_core*S0*(self.rvol_grid/self.pro_grid)[:,None],0)  # sum over radius
         self.total_source += self.src_div   # units of particles/s/cm
 
-        if self.wall_recycling>0: # recycling activated
+        if self.wall_recycling >= 0: # recycling activated
```
fsciortino, line 271: I don't think that this is right. Recycling is only really activated when `wall_recycling>0`. If `wall_recycling=0`, we have divertor "return fluxes", but no real recycling. That's the strange nomenclature/convention that we currently have, sort of following the one in STRAHL. We could change this too, but it likely wouldn't be backwards-compatible...

odstrcilt: The edge transport model is turned off when recycling_flag is False. In this case, the self.wall_recycling in aurora is set to -1. When self.wall_recycling is 0, the model should be still turned on, only no particles trapped on the wall return back in the plasma. As it was done before, for R = 0.0001 and R = 0 the results were completely different.

fsciortino:
> Another issue, if I need to vary a source during the inference, I will have to call the function 'setup_kin_profs_depts' or at least 'get_radial_source' in every iteration.

@odstrcilt regarding this, I don't think that we can do much to improve this. Depending how what kind of source option you're  interested in, I would recommend calling `get_radial_source` at every iteration, or just changing `self.src_core` via direct creation/interpolation of whatever source profile you have.

odstrcilt:
@fsciortino I have solved the issue with varying source in the IMPRAD GUI. If the source is 1D function, it will call source_utils.get_source_time_history in each iteration and multiply it with source_rad_prof in aurora class to get src_core. 
And for any source (including 2D), there is a new variable imp+'_source_scale', which will just scale asim.src_core by some number.

odstrcilt:
@fsciortino new issue, if the kinetics profiles are changing in time, particle conservation is not working.  The recycling source will have to vary in time as well.

fsciortino:
@odstrcilt OK, I updated `rcl_rad_prof` to be time-dependent. I also merged some things from master into this branch -- I'm kind of worried that there may be something wrong in `atomic.interp_atom_prof` at this point... Let me know how it looks to you.

odstrcilt:
@fsciortino I have found two issues with interpolation. If the grid was too fine, it has not detected the non-equal spacing and it was crashed during extrapolation. But else it works fine, I have compared it with RectBivariateSpline and the results are nearly exactly the same.

fsciortino:
Merging since we seem to have resolved all the raised issues. Glad that we got this done, @odstrcilt !

PR #39, opened 2021-08-13
Prevent error when somebody set zero source

odstrcilt:


odstrcilt:
Yes, it is a trivial change, but useful for users if they forget to set the source properly. I thought that the title is descriptive enough, it is just single line change.

fsciortino (approved):

Issue #40, opened 2021-11-03
Questions to interpolation in computing the fractional density in ~aurora/atomic.py


zli11:
As a senior to the Aurora, I have a few (maybe naive) questions as following in computing the fractional density of He ionization/recombination (~aurora/atomic.py) with this GREAT framework:

1.	In the following line: 

[https://github.com/fsciortino/Aurora/blob/b64a8ca08ecfe400b3e8a090ce15476536d645fd/aurora/atomic.py#L864](https://github.com/fsciortino/Aurora/blob/b64a8ca08ecfe400b3e8a090ce15476536d645fd/aurora/atomic.py#L864)
   
   I guess _abs(table-table[...,[0]]).all()_ aims to test if the dimension of X is nontrival, thus a 1D interpolation could work instead of the 2D. However the function of _.all()_ always returns a bool with _true=1_ and _false=0_ which is in the opposite of the original expectation. 

2.	For: 

[https://github.com/fsciortino/Aurora/blob/b64a8ca08ecfe400b3e8a090ce15476536d645fd/aurora/atomic.py#L867](https://github.com/fsciortino/Aurora/blob/b64a8ca08ecfe400b3e8a090ce15476536d645fd/aurora/atomic.py#L867)

I’m a bit comfused with this scale: _*np.log(10)_. Since x, y, and table here are all the indeces of 10. Does it really mean _10**np.log(10)_ which seems make no sense?

3.	The interpolation with CartesianGrid requires an equally spaced grids in the table of adf11.
[https://github.com/fsciortino/Aurora/blob/b64a8ca08ecfe400b3e8a090ce15476536d645fd/aurora/atomic.py#L690](https://github.com/fsciortino/Aurora/blob/b64a8ca08ecfe400b3e8a090ce15476536d645fd/aurora/atomic.py#L690) Unfortunately the listed logne and logTe in acd96/scd96 for He are not uniformed (i.e [https://open.adas.ac.uk/detail/adf11/acd96r/acd96r_he.dat](https://open.adas.ac.uk/detail/adf11/acd96r/acd96r_he.dat)). Is there any module can deal with that?

Looking forward to your reply!

Cheers,
Zhi

fsciortino:
Hi @zli11 , 

thank you very much for your detailed questions. I'll address them in order:

1. thanks for spotting this mistake. Indeed, the `all()` operation was in the wrong place, which probably meant that 2D interpolation was always been used (with a small slow-down). 

[...]

PR #41, opened 2021-11-03
Modifications to atomic.py

fsciortino:
Updates to atomic.py to address issue #40 .

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.7'
+        python-version: '3.9'
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
-        python-version: '3.x'
+        python-version: '3.9'
     - name: Install dependencies
       run: |
         python -m pip install -r requirements.txt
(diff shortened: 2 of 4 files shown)


zli11 (approved):
Thanks for the changes and comments. It's much more clear to me now!

fsciortino:
@zli11 hopefully this is going in the direction of addressing all your questions!

@odstrcilt I made a quick attempt at catching the cases where the ADAS grids are not exactly equally spaced (e.g. 96 acd/scd files for He). It's actually not entirely clear to me how bad CartesianGrid should perform in these cases. I'm skeptical that rewriting linear interpolation in fortran could be faster than the numpy alternatives, so I'm not going to put time into writing that code yet... Could you please have a look?

zli11:
> @zli11 hopefully this is going in the direction of addressing all your questions!
> 
> @odstrcilt I made a quick attempt at catching the cases where the ADAS grids are not exactly equally spaced (e.g. 96 acd/scd files for He). It's actually not entirely clear to me how bad CartesianGrid should perform in these cases. I'm skeptical that rewriting linear interpolation in fortran could be faster than the numpy alternatives, so I'm not going to put time into writing that code yet... Could you please have a look?

Thanks for the so quick response and most of my concerns are solved!
BTW, for this equally spaced problem, I checked with ne=5e12 and Te=2e1, the difference compared to the table is at ~1-5% (_[[25146.64230869  1003.85272985]]_ vs _[[26159.66888482   995.81582847]]_). Still a number we can stand. :)

Cheers,
Zhi

fsciortino:
thanks for checking those numbers, @zli11 ! Glad to hear that this helped :)

fsciortino:
The reason why the automated tests were not passing was a cython bug:
https://bugzilla.redhat.com/show_bug.cgi?id=1895315
Changing temporarily the testing to occur in python3.9 only has eliminated the problem.

fsciortino:
Combined with PR #37, this PR should address all the raised points/questions. Thanks @zli11 for getting in touch!

PR #42, opened 2021-11-03
Enable Aurora to run impurity transport simulations for AUG

fsciortino:
This PR is part of a new project to enable use of Aurora at AUG and also make OMFIT more accessible/useful at IPP. 

This PR begins with a simple function that allows construction of a proxy for the EFIT geqdsk dictionary built by OMFIT. I've also written an example script in the "examples" directory to illustrate how everything can work.

Further refinements may be useful/needed.
The next step in this work is to add AUG spectroscopic diagnostics to the ImpRad module in OMFIT, which uses Aurora for forward modeling. 

Many thanks to @tardini for his help with aug_sfutils!


aurora/grids_utils.py
```diff
@@ -743,17 +747,16 @@ def build_aug_geqdsk(shot, time):
     geqdsk['fluxSurfaces']['Z0'] = Z0
     geo = geqdsk['fluxSurfaces']['geo'] = {}
     
-    pf = equ.get_profile('PFL')[it]
```
odstrcilt: One small caveat CLISTE is using a different definition of poloidal flux compared to EFIT, it differs by a factor of 2pi. I think that it does not matter here, but if you need to calculate anything which depends on the absolute value of Psi, you should be aware of it.

odstrcilt (approved):

aurora/grids_utils.py
fsciortino: thanks for mentioning this, @odstrcilt . Lots of little things to be careful about... 

This should be good enough as a first implementation, although this can likely be improved with extra help from @tardini over time.

fsciortino:
Right now something in the automatic testing is failing -- it seems to be an issue on the cython side in scipy... Unless someone can help here and identify the issue, I'm going to wait before merging into master, hoping that the issue is solved upstream.

aurora/grids_utils.py
```diff
@@ -701,21 +703,23 @@ def build_aug_geqdsk(shot, time):
...
+    it = np.argmin(abs(eqm.time - time))
 
     # build up critical geqdsk components
-    geqdsk = {}
-    geqdsk['NW'] = len(equ.Rmesh)
-    geqdsk['NH'] = len(equ.Zmesh)
-    R0 = equ.Rmag[it]
-    Z0 = equ.Zmag[it]
+    from omfit_classes import omfit_eqdsk
+    geqdsk = omfit_eqdsk.OMFITeqdsk('g123456.12345') # initialized new gEQDSK
```
orso82, line 713:
```python
geqdsk = omfit_eqdsk.OMFITeqdsk(f'g{shot:06d}.{time:05d}') # initialized new gEQDSK
```

aurora/grids_utils.py
```diff
@@ -747,25 +751,42 @@ def build_aug_geqdsk(shot, time):
...
     
     geqdsk['RBBBS'], geqdsk['ZBBBS'] = r[0][0], z[0][0]
+    
+    # a few extra things to enable greater use of omfit_eqdsk
+    geqdsk.add_rhovn()
+
+    geqdsk['PRES'] = eqm.pres[it,ind]
+    geqdsk['PPRIME'] = eqm.dpres[it,ind]  # tentative
+    geqdsk['FFPRIM'] = eqm.ffp[it,ind]
+    #embed()
```
orso82, line 781: Make sure your gEQDSK is in COCOS 1, and add `AuxQuantities` and `FluxSurfaces`
```python
geqdsk._cocos=COCOS_OF_THE_QUANTITIES_THAT_YOU_ENTERED
geqdsk.cocosify(1, calcAuxQuantities=True, calcFluxSurfaces=True)
```

fsciortino: I had to play around quite a bit to get this to work -- I'm almost certain that I made up some quantities... but now `geqdsk.cocosify` runs and that's at least some progress.

PR #43, opened 2021-11-21
initial work on popcons -- close to functional

fsciortino:
Adding functionality to create Plasma OPeration CONtour (POPCON) plots, based on the work of the 22.63 MIT/Princeton class of fall 2020. Adapted Open_POPCON scripts to use atomic rates from ADAS via Aurora and Hale-Bosch reactivity already coded in `omfit_classes`.

Work in progress. 

@anelson0528 @sfrnk

### DIFF
--- a/aurora/__init__.py
+++ b/aurora/__init__.py
@@ -26,4 +26,6 @@ from .synth_diags import *
 from .solps import *
 from .kn1d import *
 
+from .popcons import *
+
 aurora_dir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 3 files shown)


fsciortino:
@anelson0528 @sfrnk There is something funny about how profiles are being treated at the moment... and I haven't had time to deal with it yet. I've added some code that uses Aurora to find profiles of impurity charge state fractional abundances, so that the average Z for a given impurity species at each value of Te can be determined. If one sets a fixed impurity concentration value and uses these fractional abundances, then Zeff will not be flat over the radius. The code needs to be updated to deal with this.

I don't understand why functions like `_P_rad` are multiplied by `rho` at the very end. It sounds like an approximation for the specific parametrization of kinetic profiles, but I'm not sure that I follow why that's a good idea...

I'd appreciate if you could have a look and make any adjustments that you see fit. Thanks!

PR #44, opened 2021-12-14
first version of facit

DanielFajardoJ:
Work in progress on neoclassical analytical code FACIT. 
See upcoming paper by D. Fajardo et al. 2021


fsciortino:
@odstrcilt The work on FACIT is effectively finished - many thanks to @DanielFajardoJ for his efforts to make this work!
A few notes / outstanding issues:
- The code in facit.py has been all benchmarked by Daniel; its results correspond to those in @DanielFajardoJ 's paper.
- The script in examples/facit_basic.py shows how to run FACIT + give an example of how to use the D's and V's from FACIT into Aurora, which is a relatively trivial operation.
- In order to run the `rotation_model=2` (needed for impurities heavier than Ar), one needs to compute the R and V values along each flux surface. Daniel and I have compressed some of @odstrcilt 's routines to make a `rhoTheta2RZ` function in coords.py, but there is clearly some issue that we cannot figure out. @odstrcilt could you please have a look? Hopefully you can see the problem, having familiarity with the logic of that code. As far as I can see, there is an issue with the use of PSIRZ from the geqdsk dictionary, but I cannot see what. In the `mapeq.py` version for AUG shotfiles, the poloidal flux matrix (`eqm.pfm`) is used. This should be exactly the same as geqdsk['PSIRZ'], as far as I can tell... but somehow there's still an issue.
- we may want to clean up the examples/facit_basic.py script a bit more.

Once the issue with rhoTheta2RZ is solved, we should be ready to merge, I think.

odstrcilt:
The source code looks fine, I will try to run the example tomorrow and hopefully, it will work.

examples/facit_basic.py
```diff
@@ -0,0 +1,240 @@
...
+            gradNz = np.gradient(Nz, roa*amin)
+
+            fct = aurora.FACIT(roa,\
+                        zi, asim.A_imp,\
+                        asim.main_ion_Z, asim.main_ion_A,\
+                        Ti, Ni, Nz, Machi, Zeff, \
+                        gradTi, gradNi, gradNz, \
+                        amin/R0, B0, R0, qmag, \
+                        rotation_model = rotation_model, Te_Ti = TeovTi,\
+                        RV = RV, ZV = ZV) #JV = JV)
```
odstrcilt, line 191: why is it not passing the JV argument?

aurora/facit.py
```diff
@@ -0,0 +1,1461 @@
...
+    Dz_* : 1D array (nr)
+        diffusion coefficient for each flux component [:math:`m^2/s`]
+        * = PS, BP or CL (for Pfirsch-Schlüter, Banana-Plateau and Classical)
+    Kz_* : 1D array (nr)
+        coefficient of the main ion density gradient for each flux component [:math:`m^2/s`]
+    Hz_* : 1D array (nr)
+        coefficient of the main ion temperature gradient for each flux component [:math:`m^2/s`]
+    Vrz_* : 1D array (nr)
+        radial flux per particle (Flux_z/nz)_* for each flux component [:math:`m/s`]
+    Vrz_* : 1D array (nr)
```
odstrcilt, line 150: Vrz_ is here twice

fsciortino: I suspect that the second one should be deleted; there doesn't seem to be a variable like Vconv_* fr each flux component. @DanielFajardoJ could you please confirm and make the correction?

fsciortino: Also, I think that we way want to substitute the header "Returns" with "Returns (as attributes)". The __init__() method here does not actually use a return statement.

fsciortino:
@odstrcilt did you get a chance to do some more testing? No worries if you need some time - @DanielFajardoJ is away for some time anyway, but it would be nice if we could conclude as soon as he's back.

odstrcilt:
I have just tested that it works, but I have not compared the results with NEO by myself. I have read Daniel's paper. 
Once it will be in AURORA, I would like to use FACIT to calculate the W transport coefficient in the postprocessing step of OMFIT profiles, and I will add it also to the NEO module such that it will always be plotted together with NEO results.

aurora/facit.py
DanielFajardoJ: The second Vrz_* was wrong, it should have been Vz_* for the convection of each PS, BP, CL flux component which were the attribute names. I now changed these names to Vconv_* for consistency and updated the header (also the "Returns (as attributes)").

DanielFajardoJ:
Thank you both @fsciortino and @odstrcilt for all the support.

fsciortino (approved):
Thanks @DanielFajardoJ ! We discussed this with @odstrcilt and we think that it's ready! Thanks for all your work. Time to integrate FACIT in other workflows with Aurora :)

Issue #48, opened 2022-01-11
Identical wavelength for different transitions in the same file


Didou09:
Motivations:
--------------
* In its current version, aurora load spectral lines from `ADAS` `adf15` files into a dictionary where they are sorted by wavelength (i.e.: the wavelentgth, as a float, is the key of the dictionary)
* This breaks down if, unfortunately, several different transitions share the same wavelength.
* A long-term fix would be to set a unique key to each transition, using something else than the wavelength.

Current behaviour:
---------------------
On python `3.7.11`, the behaviour when fed several times the same key is that the dict will only retain the last provided value:
```
In [1]: {0: 1, 0: 2}
Out[1]: {0: 2}
```
It means that in its current version, `aurora`, when confronted with several transitions of identical wavelengths, will only provide the one it encounters last in the file. 

Example:
----------
* Interestingly, it happens for` Ar16+`, when ones looks for the `k` lines, it turns out a different transition with the same wavelength (`3.9900 A`) exists in the same file `pec#ar15.dat`:
![image](https://user-images.githubusercontent.com/5929562/148978656-b53fa19d-cd5c-4aab-9da1-7980be48c4c7.png)

Didou09:
Hi @fsciortino , thanks for the update on master.

Using it raised one error, that I corrected on my fork of aurora, and then a second error occured, that I also corrected on my fork.

Then it works, but I started thinking that when a line wavelength is shifted, the  code should raise a warning so the user knows it, so I implemented the warning.

Using that version of the code revealed that there are actually many lines with redundant wavelengths.

So I implemented a more synthetic warning that is raised only once and gives the list of all redundant wavelengths (ordered according to the wavelength).

```
In [11]: import aurora

In [12]: filepath_li = '/Home/DV226270/Diag_XICS/Atomic_data/Aurora/pec#ar15.dat'

In [13]: di = aurora.read_adf15(filepath_li)
```

![image](https://user-images.githubusercontent.com/5929562/149133826-9c6ca7dd-539a-425f-9cf8-cc487268701f.png)


For file here is the result: 183 lines have redundant wavelength.
I see at least 2 causes for concern:
1. you will notice that more than 2 transitions can share the same wavelength
   => shifting the wavelength by `+1e-6 A` is not enough, because when 3 lines have the same wavelength, 2 of them will remain redundant (both shifted by `1e-6 A`, so still equal)
  => Incremental shifts may fix that
[...]

PR #49, opened 2022-01-12
Redundant lines: 2 bug fixes + synthetic warning

Didou09:
Motivations:
--------------
* Bugfix to set dict keys to ADAS indexing was not working properly when index_lines=False because:
    * `isel` was not defined (but used)
    * dict key was set priori to shifting
* Lines with redundant wavelength in a `adf15` file seem to be more common than orignally thought and several uneasy cases can occur.

Main proposed changes:
----------------------------
* The 2 bugs are fixed
* A synthetic warning is raised when lines are shifted due to redundancy, it lists all lines and wavelength and makes visible some uneasy cases with mulitple redundancies and/or unfortunate discrepancies exactly equal to the implemented shift by `+1e-6 A`.
* See [here](https://github.com/fsciortino/Aurora/issues/48#issuecomment-1010965722) for details and thoughts on possible solutions.

Feel free to merge or just to use as back-up.


fsciortino (approved):

Didou09:
OK, do you want for future PRs, that I push on a dedicated branch in your repo instead of doing it via my fork ?
This way you would be able to push to said branch.

If that solution is better for you you need to make sure to include me as a collaborator in your settings so I can actually push to your repo direcly, on non-critical branches.

fsciortino:
just added you as collaborator 👍

PR #51, opened 2022-03-16
Provide parsed ADF15 files in the form of pandas DataFrames

fsciortino:
Significant changes to the `aurora.radiation.read_adf15` function. This is NOT a backwards-compatible change and will therefore require that all Aurora users are made aware of the change and of the nice new capabilities that are now available. 

I copy here below a section of the updated docs for the `read_adf15` function, useful for testing:

```
    To plot the Lyman-alpha photon emissivity coefficients for H (or its isotopes), one can use:

    >>> import aurora
    >>> filename = 'pec96#h_pju#h0.dat'
    >>> # fetch file automatically, locally, from AURORA_ADAS_DIR, or directly from the web:
    >>> path = aurora.get_adas_file_loc(filename, filetype='adf15')  
    >>> # load all transitions provided in the chosen ADF15 file:
    >>> trs = aurora.read_adf15(path)
    >>> # select the excitation-driven component of the Lyman-alpha transition:
    >>> tr = trs[(trs['lambda [A]']==1215.2) & (trs['type']=='excit')]
    >>> # now plot the rates:
    >>> aurora.plot_pec(tr)

    Note that excitation-, recombination- and charge-exchange driven components can
    be fetched in the same way by specifying the `type`, e.g. using
    >>> trs['type']=='excit' # 'excit', 'recom' or 'chexc'

    Since the output of `aurora.radiation.read_adf15` is a pandas DataFrame, its contents
    can be indexed in a variety of ways, e.g. one can select a line based on the 
    ADAS "ISEL" indices and plot it using
    >>> aurora.plot_pec(trs.loc[(trs['isel']==1)])
    or simply
    >>> aurora.plot_pec(trs.iloc[0])
    Spectral lines are ordered by ISEL numbers -1 (Python indexing!), so using the 
    `iloc` method of a pandas DataFrame allows effective indexing by ISEL numbers.

    The log-10 of the PEC interpolation function can be used as
    >>> tr['log10 PEC fun'].iloc[0].ev(np.log10(ne_cm3), np.log10(Te_eV))
    where the interpolant was evaluated (via the `ev` method) at specific points of `n_e`
    (units of [:math:`cm^{-3}`]) and `T_e` (units of :math:`eV`). Note that the log-10
    of `n_e` and `T_e` is needed, not `n_e` and `T_e` themselves!

    Metastable-resolved files are automatically identified based on the file nomenclature
    and parsed accordingly, e.g.::

    >>> filename = 'pec96#he_pjr#he0.dat'
    >>> path = aurora.get_adas_file_loc(filename, filetype='adf15')
    >>> trs = aurora.read_adf15(path)
    >>> aurora.plot_pec(trs[(trs['lambda [A]'==584.4])])
```

I would definitely appreciate if reviewers could test these changes on a number of other ADF15 files. I've tried to make the new methods compatible with a number of ADF15 formats (yes, there are many...), but I may have missed something!

### DIFF
--- a/aurora/radiation.py
+++ b/aurora/radiation.py
@@ -31,12 +31,12 @@ from scipy import constants
 import warnings, copy
 from scipy import constants
 from collections import OrderedDict
+import pandas as pd
 
 from . import atomic
 from . import adas_files
 from . import plot_tools
-
-
+    
 def compute_rad(
     imp,
     nz,
@@ -817,11 +817,6 @@ def parse_adf15_configs(path):
     If this function fails on one of your files, please report this via a Github 
     issue and one of the Aurora core developers will help adapting the parser.
     '''
-    try:
-        import pandas as pd
-    except ImportError:
-        raise ImportError('Missing installation of pandas!')
-    
     with open(path, "r") as f:
         lines = f.readlines()
         
@@ -881,11 +876,6 @@ def parse_adf15_spec(lines, num_lines):
     dict : 
         Dictionary containing information about all loaded transitions.
     '''
-    try:
-        import pandas as pd
-    except ImportError:
-        raise ImportError('Missing installation of pandas!')
-
     # collect info on transitions        
     while True:
         l = lines.pop(0)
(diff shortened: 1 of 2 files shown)


TGleiter (approved):
Works very nice! :)

fsciortino:
@odstrcilt @aarmros @Didou09 I would appreciate if you could have a look at this PR before it is merged because **it will affect any usage of `aurora.read_adf15` in your code**. Other users will also be impacted by this change that is *not* backwards-compatible. As I wrote above, it seems worth doing this, because the new capability can help in a number of interesting applications. 

@odstrcilt the ability to select lines from the new "relational database" using pandas will affect the selection of lines from a selected ADF15 file in ImpRad, when trying to match SPRED signals.

@aarmros the way you have been selecting Ly-a and C lines will need updating (see notes above)

@Didou09 using the pandas' DataFrames that are now returned by `aurora.read_adf15`, it is much easier/nicer to select all lines within a certain wavelength region (e.g. when looking at the He-like Ar spectrum) or select only lines populated by excitation, etc..

Thanks in advance for your feedback ;)

odstrcilt:
Hi Francesco,
could you also write a function that will convert the panda object back to
the dictionary of dictionaries of liens, as it was before?

fsciortino:
Hi @odstrcilt , 

if you load an ADF15 file using 
```
import aurora
filename = 'pec96#h_pju#h0.dat'
path = aurora.get_adas_file_loc(filename, filetype='adf15')  
trs = aurora.read_adf15(path)
```
you can do 
```
trs.to_dict()
```
and this will be a standard dictionary, but significantly "augmented" wrt what was possible before this PR. You should then get the following:
```
>>> trs.to_dict().keys()                                                                                            
dict_keys(['isel', 'lambda [A]', 'transition', 'type', 'log10 PEC fun', 'dens pnts', 'temp pnts', 'PEC pnts'])
```
You can also use `to_xarray()` if you prefer to work with xarray structures. I think that you'll agree that this is a much better functionality than previously available with Aurora. But, yes, it is not entirely backwards compatible and requires users to do some small adaptations to their code. I think that Aurora is too young as a package to already be stuck in backwards-compatibility issues :) Better to do this kind of changes now than in 2 years.

It didn't take me long to adapt my own code. I suggest that you give it a try and you'll see quickly the nice benefits for spectroscopy work ;)

fsciortino:
Planning to merge by the end of the day. I've been told by @aarmros that this PR worked fine for him. 

If anyone else would like to comment or ask questions before we merge, please do so now.

inwoo-song:
Hi Francesco,

 When I read ADF15 from open-ADAS, there sometimes have different formats. The header of each block in the PEC files sometimes has no blank space so I have trouble to find “/type=excit” thing. But I fixed it by adding the following in radation.py. Could you also check this out?

below the 690th line where l=lines.pop(0), add

        l = l.replace('/', ' /').replace('=', ' = ')    
        if l.find('TYPE') == -1:
            l = l.replace('type', 'TYPE')    

image

fsciortino:
thanks @inwoo-song for bringing this up! Actually, while talking to @TGleiter we also saw that "TYPE" is not always in capital letters in ADF15 files. Easy to fix! 

There are likely other "exceptions" to the formatting that are not yet caught by the Aurora parser. Whenever you encounter these, it'd be great if you could open a new Issue. Thank you!

PR #52, opened 2022-03-25
small bug in dimensions of rcl_rad_prof fixed

TGleiter:
Small problem with the missing time dimension of rcl_rad_prof in core.py, setup_kin_profs_depts fixed by adding a np.broadcast.


fsciortino (changes requested):
thanks @TGleiter ! 
Right now, the np.broadcast command is repeated twice, in both the "if" and the "else". Could you please modify this to have it only once (out of the if statement)?

fsciortino:
thanks @TGleiter !

fsciortino (approved):

PR #54, opened 2022-04-07
New imp brems

aarmros:
Altered impurity_brems calculation to allow calculation of bremsstrahlung at specific wavelengths or integrated over a specified range of wavelengths. The function still behaves the same to the end user but has additional options which allows the new operations.


fsciortino (approved):
Looks good! Thanks @aarmros for taking the time to share this update :)

PS: next time please remember to start from the latest version of the `master` branch, or use `git merge master` to merge `master` into your "feature branch" (`new_imp_brems`, in this case) before opening a PR.

PR #55, opened 2022-04-25
New branch ab

abhatt03:
a59576bfcb5d17fdf1593c44831f9178d4ed9942

### DIFF
--- a/aurora/atomic.py
+++ b/aurora/atomic.py
@@ -1011,7 +1011,7 @@ def gff_mean(Z, Te):
 
     # set min Te here to 10 eV, because the grid above does not extend to lower temperatures
     Te = np.maximum(Te, 10.0)
-
+    Z  = np.ravel(Z)
     log_gamma2 = np.log10(Z ** 2 * thirteenpointsix / Te)
 
     # dangerous/inaccurate extrapolation...
--- a/aurora/atomic.py
+++ b/aurora/atomic.py
@@ -1071,7 +1071,7 @@ def impurity_brems(nz, ne, Te, freq='all', cutoff=0.1):
     # plasma frequency (divide by 2pi to have units of Hz)
     fp = np.sqrt(1e6*ne*constants.e**2/(constants.epsilon_0*constants.m_e))/(2.*np.pi)
 
-    freqT = np.tile(freq, (ne.shape[1], ne.shape[0], 1)).T
+    freqT = np.tile(freq, (ne.shape[0], 1)).T
 
     # find all values below fw and above cut
     mask = np.logical_or(freqT<fp, freqT>cut)
@@ -1093,7 +1093,8 @@ def impurity_brems(nz, ne, Te, freq='all', cutoff=0.1):
         intgrnd = np.exp(-constants.h*freqT/(Te*constants.e))
         intgrnd[mask] = 0
         intV = simps(intgrnd, freqT, axis=0)
-        
+
+    Z = np.ravel(Z)    
     return 4*np.pi*Z**2*nz[:,1:]*gff*(ne*1e12*const*np.sqrt(1/(Te*constants.e))*intV)[:,None]
 
 
--- a/aurora/core.py
+++ b/aurora/core.py
@@ -1096,7 +1096,8 @@ class aurora_sim:
         Z = np.arange(Zmax + 1)
         self.delta_Zeff = nz * (Z * (Z - 1))[None, :, None]  # for each charge state
         self.delta_Zeff /= self.ne.T[:, None, :]
-
+        return self.delta_Zeff
+        
     def plot_resolutions(self):
         """Convenience function to show time and spatial resolution in Aurora simulation setup.
         """
--- a/aurora/interp.py
+++ b/aurora/interp.py
@@ -109,7 +109,7 @@ def ratfun(params, d, rLCFS, r):
 def interp_quad(x, y, d, rLCFS, r):
     """Function 'interp' used for kinetic profiles.
     """
-    f = interp1d(x, np.log(y), kind="quadratic", assume_sorted=True, copy=False)
+    f = interp1d(x, np.log(y), kind="quadratic", assume_sorted=True, copy=False,fill_value = 'extrapolate')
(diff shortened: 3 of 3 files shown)


fsciortino:
@abhatt03 thanks for opening a PR, but please add a description of what you're trying to do, what issues you met and why you are proposing to solve them the way you are.

abhatt03:
@fsciortino : 2) commit id 2f1be45: The original x-axis fed to the Aurora namelist in my case is in open interval (0,1). However, rho_p and rho_vol grid in Aurora covers the range [0,>1]. In such cases (in general), without the feature fill_value = 'extrapolate' using scipy.interpolate.interp1d will generate an error (as happens here and also something I have encountered before). As a result, added the feature fill_value = 'extrapolate' to line 112 and line 131  in interp.py

abhatt03:
@fsciortino : 3) commits fba3c63: Without (new) line 1014: Z  = np.ravel(Z), the function (as the error describes) is trying to project a tuple (Zeff i.e., Z) of size (1,x,1) onto a tuple of (y,1) while conducting the calculation in (new) line 1015: log_gamma2 = np.log10(Z ** 2 * thirteenpointsix / Te). The statement np.ravel changes the Z.size from (1,x,1) to (x,) hence making the calculation in line 1015 error free.

abhatt03:
@fsciortino : 4) commit id fbbd0d9 : a) line 1097 is the same change in Z (to change Z.size) as commit fba3c63
b) ne provided to Aurora namelist (at least in my case) is an array. So ne.shape = say (x,). ne.shape[1] in (earlier) line 1074: freqT = np.tile(freq, (ne.shape[1], ne.shape[0], 1)).T,  in atomic.py , generates an error. Hence this line in atomic.py is modified to: freqT = np.tile(freq, (ne.shape[0], 1)).T

aurora/atomic.py
```diff
@@ -1071,7 +1071,7 @@ def impurity_brems(nz, ne, Te, freq='all', cutoff=0.1):
     # plasma frequency (divide by 2pi to have units of Hz)
     fp = np.sqrt(1e6*ne*constants.e**2/(constants.epsilon_0*constants.m_e))/(2.*np.pi)
 
-    freqT = np.tile(freq, (ne.shape[1], ne.shape[0], 1)).T
+    freqT = np.tile(freq, (ne.shape[0], 1)).T
```
fsciortino, line 1074: @abhatt03 thanks for suggesting these changes. However, they don't seem to be in agreement with the array shapes expected by this function, as described in the function docsstring. For example, `ne` is expected to be both a function of space and time. If you are trying to use this function with `ne` being only a function of radius, I'd suggest giving `ne[None]` (or similar) as input, rather than just `ne` -- then it should work. The `np.ravel(Z)` modification above seems related to this same matter of including time dependence in array shapes.

aurora/core.py
```diff
@@ -1096,7 +1096,8 @@ def calc_Zeff(self):
         Z = np.arange(Zmax + 1)
         self.delta_Zeff = nz * (Z * (Z - 1))[None, :, None]  # for each charge state
         self.delta_Zeff /= self.ne.T[:, None, :]
-
+        return self.delta_Zeff
```
fsciortino, line 1099: @abhatt03 the `return` statement was intentionally avoided here, because `delta_Zeff` is anyway cached in `self`. This means that you can anyway access the result of your `aurora_sim` instantiation (say, called `asim`), as `asim.delta_Zeff`.

abhatt03: @fsciortino : I am attaching three screenshots to show the reason why I made the changes. Forgive me if the screenshots are too green. In the Aurora manual, as I read before starting to work with Aurora, the function to obtain the Zeff values is calc_Zeff() in core.py (Zeff values obtained with asim.calc_Zeff()). 

I am attaching the screenshots, where, two of the screenshots show error received with asim.delta_Zeff() (it may be helpful to mention this is the quantity and not calc_Zeff() in the manual if a return statement is not desired) and asim.calc_Zeff(), when the return statement is commented. The third one shows what happens when the return statement is uncommented. 

cal_zeff_error_received_Aurora1

cal_zeff_error_received_Aurora2

cal_zeff_error_corrected_received_Aurora1

As these errors are what I have received, whatever changes I have mentioned until now are simply stated/intended as a suggestion.

aurora/atomic.py
abhatt03: @fsciortino : So I took your suggestion and I began declaring all the arrays in the Aurora namelist as Y[:,None] say for e.g., ne. I am attaching a screenshot to describe the error I receive in such case.

Error_ne_dimension_Aurora1

abhatt03:
@odstrcilt : Thank you for the suggestion. Sorry, I did not see your comment earlier and went ahead with posting the new screenshots (updated my comment) based on my earlier corrections. I will definitely take a look again and apply the new changes you suggested.

abhatt03:
Hello Francesco,

Thanks very much for reaching out!
I solved the PR issue in my local version (just tweaked a few lines as
I had mentioned there as well so that my script works). So this PR can be
removed.

I have a completely different issue this time. I am attaching a figure (one
of Dz and one of Vz) for a shot I am analyzing. Don't be panicked by
the magnitudes (I was!) but as you will see I am getting a magnitude
almost 10^3 higher than expected. I was wondering if by any chance either
ionization (scd) or recombination (acd) term (~10^6 in cm3/s to m3/s or
vice versa) fed to the impurity transport equation in Aurora from the ADAS
database is changing somewhere by any chance or if there is a volume term
from cm^3 to m3 or vice versa (Please don't get me wrong, this is just my
extreme attempt to understand these values). I have already checked in
impden0 and impden1 - I believe scd and acd is extrapolated in log10.
Still, if by any chance you could let me know something in this regard, it
would be really very helpful.

Thanks very much and hope all is going well with you and your research.

Best,
Amrita

abhatt03:
@fsciortino : I have it solved in the local version I am using. Thanks Francesco.
Amrita

fsciortino:
OK, thanks for the clarification @abhatt03 !

PR #59, opened 2022-07-28
fix to rescale n0 in ImpRad

fsciortino:
Quick fix for OMFIT ImpRad framework to freely rescale neutral density (n0) within inferences of impurity transport -- requested by @odstrcilt . 

mention @TGleiter

### DIFF
--- a/aurora/core.py
+++ b/aurora/core.py
@@ -448,9 +448,13 @@ class aurora_sim:
             include_cx=self.namelist["cxr_flag"],
         )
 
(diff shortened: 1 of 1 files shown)


odstrcilt (approved):
as it was done before, the value of self.alpha_RDR_rates would not be equal to original Rne, but Rne+ alpha_CX_rates

fsciortino:
Thanks for checking this, @odstrcilt !

PR #62, opened 2022-10-05
Corrected bugs in the particle conservation

antonellozito:
Corrected bugs affecting the particle conservation:

- In **core.py**, in the calculation of the total number of impurity particles in the plasma, the volume integral should be extended over the entire radial grid and not only until the LCFS (i.e. using `rvol_max = None` rather than `rvol_max = asim.rvol_lcfs`) in order to really account for all the particles.
- In **main.f90** corrected a bug (present only when the recycling is deactivated), in the **edge_model** subroutine, which prevents the removal of particles from the divertor reservoir through pumping, leading to a continuous accumulation of particles in there.


fsciortino (approved):
Checked in person with @antonellozito , also included in `recycling_model` branch PR.

PR #65, opened 2022-10-17
Enabling Aurora on W7-X

fsciortino:
Made a few changes to core.py to allow users to pass some key parameters for the grid directly through the namelist. Everything else works fine, except that no use of geqdsk files is possible for W7-X/stellarators, so all Aurora functions/methods invoking geqdsk processing won't work. This does not prevent one from running single Aurora simulations.

Users will need to pass a minor radius and a major radius (and possibly a magnetic field strength for the prompt redeposition model), usually taking these from a VMEC reconstruction. These numbers could also be set arbitrarily. These changes therefore add flexibility to Aurora simulations, permitting them to run again also without a true equilibrium.

This PR enables further development within the OMFIT ImpRad module for W7-X.

Interested parties: @thromba

### DIFF
--- a/aurora/core.py
+++ b/aurora/core.py
@@ -94,8 +94,17 @@ class aurora_sim:
         else:
             self.geqdsk = geqdsk
 
-        self.Raxis_cm = self.geqdsk["RMAXIS"] * 100.0  # cm
-        self.namelist["Baxis"] = self.geqdsk["BCENTR"]
+        if 'Raxis_cm' in self.namelist:
+            self.Raxis_cm = self.namelist['Raxis_cm'] # cm
+        elif 'RMAXIS' in self.geqdsk:
+            self.Raxis_cm = self.geqdsk["RMAXIS"] * 100.0  # cm
+
+        if 'BCENTR' in self.geqdsk:
+            self.namelist['Baxis'] = self.geqdsk["BCENTR"]
+        if ('prompt_redep_flag' in self.namelist and self.namelist['prompt_redep_flag'])\
+           and not hasattr(self, 'Baxis'):
+            # need magnetic field to model prompt redeposition
+            raise ValueError('Missing magnetic field on axis! Please define this in the namelist')
 
         # specify which atomic data files should be used -- use defaults unless user specified in namelist
         atom_files = {}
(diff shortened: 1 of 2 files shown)


fsciortino:
@odstrcilt I'd appreciate a quick review if possible, even though I assume that you don't have access to W7-X data. It would be sufficient if you could check to make sure that these changes are backwards-compatible, as they should be. You can easily test by commenting out some lines of the new w7x_example.py script and giving some reasonable values of rmin and Rmaj.

aurora/core.py
```diff
@@ -169,29 +178,42 @@ def load_dict(self, aurora_dict):
...
-            self.geqdsk, rho_pol=self.rhop_grid
-        )
+            # Save R on LFS and HFS
+            #self.Rhfs, self.Rlfs = grids_utils.get_HFS_LFS(
+            #    self.geqdsk, rho_pol=self.rhop_grid
+            #)
+            
+        else:
+            # use rho_vol = rvol/rvol_lcfs
+            self.rhop_grid = self.rvol_grid / self.rvol_lcfs
```
odstrcilt, line 214: in principle, we could remove rhop entirely from aurora and use rhov instead. It might be confusing to call it rhop even if it is something else.

aurora/core.py
```diff
@@ -94,8 +94,17 @@ def __init__(self, namelist, geqdsk=None):
...
-        self.Raxis_cm = self.geqdsk["RMAXIS"] * 100.0  # cm
-        self.namelist["Baxis"] = self.geqdsk["BCENTR"]
+        if 'Raxis_cm' in self.namelist:
+            self.Raxis_cm = self.namelist['Raxis_cm'] # cm
+        elif 'RMAXIS' in self.geqdsk:
+            self.Raxis_cm = self.geqdsk["RMAXIS"] * 100.0  # cm
+
+        if 'BCENTR' in self.geqdsk:
+            self.namelist['Baxis'] = self.geqdsk["BCENTR"]
+        if ('prompt_redep_flag' in self.namelist and self.namelist['prompt_redep_flag'])\
```
odstrcilt, line 104: prompt_redep_flag is really working in AURORA?

odstrcilt (approved):
It seems fine, I have not found anything serious.

aurora/core.py
fsciortino: Yes, one could use `rhov` or even just a general `rho` (which I don't like much though, because it leaves ambiguity. However, the issue is that this would not be a backwards compatible change... and I'm not sure that we want to make a big thing out of this change, which is only important for W7-X at the moment... so I'm tempted to just leave it like this and just "alert" W7-X folks that Aurora refers to `rhop`, but in W7-X's case this is just a rhov. (FYI, W7-X people don't talk about `rhov` either - it's just called `reff` for stellarator research, and it's the standard radial coordinate)

fsciortino:
I'm tempted to delete lines 81-93 in core.py, which allow users to automatically fetch a geqdsk from EFIT01. I don't think that we ever use this code (EFIT01 is a bad choice usually, anyway) and maybe users should be discouraged from using this. Eliminating this would also avoid confusion for users that don't want to use a geqdsk at all (e.g. on W7-X). If we delete this, then `geqdsk=None` could be used to simply indicate that key grid parameters should be searched for in the namelist. @odstrcilt do you see any reason not to do this?

fsciortino:
@odstrcilt well, EFIT01 is not MSE-constrained on DIII-D, for example. On C-Mod, EFIT01 is never the right one to do detailed analysis.

Since you've never used it either, I'll eliminate this default geqdsk option.

PR #66, opened 2022-10-27
Remove Julia Stuff (temporarily)

ajcav:
Hey Francesco, I wanted to play around with this a bit more but to start I think it should be removed until it's functional.

Hope you're well,
Amelia~


fsciortino:
Hey @ajcav , I hope you're doing well!
Thanks for your taking time to look into this. Honestly, I'd rather look for someone to fix this than remove the code. Keeping it there signals our interest in having some Julia support (which I really like). I'd be your greatest fan if you were able to fix things yourself!

PR #67, opened 2022-11-07
Fix bug in plot_2d_overview in the OEDGE interface

CorvidCanine:
The call to `oedge_output.plot_2d` in `oedge_output.plot_2d_overview` has the wrong keyword argument name for the colour bar label.
It was passing `label` instead of `cbar_label`, which throws an exception.


fsciortino:
Thanks, @CorvidCanine ! Nice catch.

fsciortino (approved):

PR #69, opened 2022-12-11
Line ratios tools

fsciortino:
General tools to examine spectroscopic line ratios using ADAS data. All the new relevant code is in a single file, lr_tools.py. The implementation resembles the one described by S. Henderson's NII line ratio papers. The code maintains a good degree of generality by asking users to specify as input for each line to be considered (a) the ADAS "ISEL" indices for excitation and recombination components (specific to a chosen ADF15 file), (b) values, and (c) uncertainties of line intensities. An arbitrary number of ISEL numbers can be considered/summed for each experimental line shape; more than one transition can be included by providing additional ISEL numbers within the "list of lists" of the 'isel' sub-dictionary for a given line. If no recombination component should be included, users should indicate "None" in the second element of each sub-list.

Example: evaluate ne and Te from the ratio of the NII line intensities at 4041A and 3995A, considering experimental uncertainties. 
```
    data = {
        'N_1_4041': {
            'isel': [[21,71]],
            'val': val_4041,
            'unc': unc_4041
            },
        'N_1_3995': {
            'isel': [[15,65]],
            'val': val_3995,
            'unc': unc_3995,
            }
        }
    atom_data = aurora.atomic.get_atom_data('N', ["scd", "acd"])
    ne, ne_unc, Te, Te_unc = get_line_ratio_ne_Te(data, cs=1, atom_data=atom_data)
```

Interested parties: @wintersv19 , F. Henke (github username?)

@ Victoria, Frederik: could you please try this out and let me know if everything works well for you, before we merge? Requests and suggestions for changes would be certainly appreciated!

### DIFF
--- a/aurora/lrtools.py
+++ b/aurora/lrtools.py
@@ -1,0 +1,304 @@
+'''Tools to analyze spectroscopic line ratios.
+'''
+# MIT License
+#
+# Copyright (c) 2021 Francesco Sciortino
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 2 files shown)


fsciortino:
@odstrcilt I made these scripts a long time ago for line ratio studies, and showed them to the groups in Greifswald and Garching, but did not hear whether it's being used or not. It's quite simple, but has some decent generality. Would you mind having a look?

aurora/lrtools.py
```diff
@@ -0,0 +1,304 @@
...
+    plot_line_ratio_space('N',1, exc4042, exc3995)
+    '''
+    # ne and Te grids  
+    ne_grid = np.linspace(min_ne, max_ne, 300)
+    Te_grid = np.linspace(min_Te, max_Te, 400)
+    NE, TE = np.meshgrid(ne_grid, Te_grid)
+
+    # fractional abundances at ionization equilibrium
+    atom_data = atomic.get_atom_data(imp, ["scd", "acd"])
+    _Te, fz = atomic.get_frac_abundances(
```
odstrcilt, line 273: Most of the code here should be possible to replace by two calls of class `tec `class

PR #71, opened 2022-12-29
Include T.  Nishizawa's steady state solver

odstrcilt:


odstrcilt:
@fsciortino I have finally implemented Takashi's model in aurora. The most difficult part was that he was using Dirichlet boundary conditions (i.e., you need to provide impurity density at the edge). I have modified it to use the same decay length condition as is in STRAHL. And it uses the same source profile as AURORA, the original solution was valid only in source free region.  Finally, I have included the option to use Z-dependent transport coefficients. 

Here is compared steady-state solution from the analytical solver and from the iterative solution by aurora:
aurora

Usually, it matches within 20%. When I set the SOL Mach number to zero, the match is almost perfect. Given all uncertainties and simplification in the edge model, this agreement is acceptable.

The solution time is not as good as I was hoping for. I have done my best to write fast python code. It can certainly be done faster in fortran, but I'm not doing it. The solution time for the new and old steady-state solvers are comparable for smaller radial grid sizes and low-Z impurities (about 50-100ms) But the computation complexity increases as O(nr^2*Z^2) (or even faster) while the original AURORA has a linear complexity with radius and impurity charge.

odstrcilt:
This code has one large advantage, it is really simple. For example, it should not be difficult to add metastables. I would have to modify adas reading routines for metastable resolved files and slightly modify the steady state solver.

fsciortino:
All tested and looking great! Thanks @odstrcilt !

fsciortino (approved):

PR #72, opened 2023-01-03
corrected error in calculation of neutral density from lyman-alpha

aarmros:
Small update in Lya_to_neut_dens which contained the wrong expression for N1 using the adas ADF15 rate coefficients. 

Previous expression
N1 = emiss_prof / E_21 / (ne * pec_exc + ni * pec_recomb)
change to 
N1 = ((emiss_prof / E_21)- ni * ne*pec_recomb) / (ne * pec_exc)

The overall effect of this error is small except in situations near the wall or diverter.


fsciortino:
Thanks, @aarmros ! Great catch.

fsciortino (approved):

PR #73, opened 2023-01-04
Generalize analytic solver for metastables

odstrcilt:
Example of metastate resolved carbon density profile compared with unresolved data 
resolved_c_profile

Example of main carbon lines evaluated using resolved and unresolved atomic data
resolved_c_lines


fsciortino:
Lots of amazing work @odstrcilt ! Thank you for putting this together. I did some tests, corrected some formatting, and it seems ready to go.

PR #75, opened 2023-01-17
Recycling + advanced PWI model for merge with master

antonellozito:
@fsciortino @odstrcilt In view of the newest abundant updates to the master version since when I completed my model (which would create a lot of conflict), I have started implementing again my additions - recycling model + advanced plasma-wall interaction model - manually to the current master version, in order to be able to perform a clean and painless merge when I will have completed it. In doing so, according to your suggestions, I will leave out the additions I have used for the transport model in my investigation (interpolations of transport profiles and ELM model), maybe leaving them for a successive pull request, and I will introduce the PWI model as a new class.
I would leave the previous branch (recycling_model) alive for now, as it is the version which, albeit being not ready for public release, has worked so far and which includes all the stuff that I used for my personal investigation.


odstrcilt:
Hi @antonellozito,
 do you think that your model can describe the saturation of walls by non-recycling gas injection? A typical example is nitrogen seeding. First, nitrogen behaves as a non-recycling impurity, but once the wall saturates, it behaves like a noble gas.

antonellozito:
> Hi @antonellozito, do you think that your model can describe the saturation of walls by non-recycling gas injection? A typical example is nitrogen seeding. First, nitrogen behaves as a non-recycling impurity, but once the wall saturates, it behaves like a noble gas.

@odstrcilt not only it can: i developed it exactly for this aim. He in W behaves in the same way. As the plasma-wall interactions as I included it rely on actual Monte Carlo calculated surface coefficients, the only free parameter for describing this effect (which must be somehow inferred from experiments) is the saturation content of an impurity into the wall. N can be modelled in the same way (although some coefficients regard the interaction of N with wall materials are still to be produced, which can still be done in the future).

odstrcilt:
Yes, I have found the right branch in your diagram. Is this R_N coefficient an input parameter or it is somehow calculated? Could I just assume that it is zero for nitrogen?             Or would it be possible to do the calculation for nitrogen (and perhaps also neon) for both W and C walls?  How do you include the specific shape and properties of the AUG wall?

fsciortino:
Hi @antonellozito , I spoke to @odstrcilt about the strategy moving forward and we agree that the way you currently implemented things is OK: we can keep a single fortran call in `run_aurora`. Can you please complete the work by adding `pwi.py`? Then, we should be ready to test everything :) We should also all collectively try to simplify the code, because at the moment `core.py` has become a little too verbose/complicated.

Reminder: please make sure that everything is backwards-compatible!

Thanks for all your work!

antonellozito:
@fsciortino @odstrcilt the updated version of Aurora containing the extended recycling model and the advanced plasma-wall interaction model is complete and working. As you suggested, with respect to the old pull request, I have left out some additions (like transport model and ELM model), but I have kept extended recycling model and the advanced plasma-wall interaction model. Now, as you suggested, when one wants to use the advanced PWI model a the new class pwi.py must be called, rather than core.py. So, the two things are kinda separate, at least on the python side. The call to the fortran routines, however, is still shared between the simple and the new model.
Because of my ignorance on how the inheritance of a class works, there might be some redundant stuff in pwi.py... hope not too many. However, it works.
The final commit also contains the updated examples (including the one for the advanced PWI model) and the (hugely) updated documentation. I also have succesfully merged the current master into it.
Also, you know it but it is worth to mention one more time: there are no backwards incompatibility, except for one: the `out` object now is not a tuple anymore, but a dictionary. All the example scripts contained in the latest commit are adapted to that, but the users need to take this into account in their personal scripts after updating. As we widely agreed, however, this is a necessary sacrifice: having an mutable object as main output of the simulation would make possible future enhancement (like this one) much easier.
I guess that it is ready for your review. Because of the major update, I took the liberty to flag this tentative version of the code as 3.0.0 (but of course we can revert it or use different numbers, if you think so).

fsciortino:
That looks great @antonellozito ! One quick thing: I see two new directories inside the `docs` directory. Are they needed for some reason?

antonellozito:
Following the latest suggestions by @fsciortino I have heavily shortened the new method `pwi.py`, avoiding to unnecessarily duplicate methods from `core.py`.
Now, at least from the point of view of the Python interface, the original model and the new extended model including a realistic description of the plasma-wall interactions for determining the recycling are really fully separated.
From the point of view of the Fortran subroutines, there are two differences of this branch wrt to the current master:
- The (optional) inclusion of new particle reservoirs, i.e. divertor wall reservoir and pump reservoir (see the pictures from the docs). This has nothing to do with the full plasma-wall interaction model! They are separate stuff. All the additions coming from this extension wrt to the current master are basically in the `edge_model` subroutine in `main.f90`, which is now a bit longer than how it was before. I stress that all my additions in this regard are optional and regulated by if statements.
- The activation of the full plasma-wall interaction model. In this case the changes are found in `impden.f90`. What is done when the PWI model is activated is that the recycling fluxes from the main wall towards the plasma, used to asume the 0th charge state profile in the transport modelling routines, are calculated by a separate subroutine `advanced_PWI_model`, which is equally called by both the algorithms `impden0` and `impden1`. 

Apart from other suggestions, everything is ready for a review.

antonellozito:
> That looks great @antonellozito ! One quick thing: I see two new directories inside the `docs` directory. Are they needed for some reason?

@fsciortino Indeed they are not. They come from my computer, I didn't notice them. I'll fix it.

odstrcilt:
@antonellozito I will review it soon. I was hoping to do it this week, but this goal is less and less likely. Hopefully next week.

fsciortino:
@odstrcilt @antonellozito it would be a pity to leave this PR hanging like this... @odstrcilt would you be willing to do the review, and @antonellozito to address any comments on a short time scale?

antonellozito:
> @odstrcilt @antonellozito it would be a pity to leave this PR hanging like this... @odstrcilt would you be willing to do the review, and @antonellozito to address any comments on a short time scale?

@fsciortino @odstrcilt  I have not worked on this branch since March, when I do my personal simulations I still use my personal version (because we removed the ELM model from this branch). However I remember that the work on this was concluded, i.e. the additional reservoirs are included in the exact same way they are in my personal version, but in a more proper way :) and with the importan addition of a whole renewed manual ;) In other words yes, it only needs to be reviewed I guess.

odstrcilt:
I have not found the time and courage to do the review. It is a considerable change to the original code...

antonellozito:
@odstrcilt @fsciortino This will be a long comment, but I'll try to summarize here once again all my changes to the code, in order to make the review process as painless as possible. You might print this and keep it as a quick reference to always have in front of you during the review ;) I am sure this would be more practical for you, then just scrolling all the past comment. Of course, I am avaialble for any clarification about both physics and technical/coding aspects.

Note that this version of the code is, locally, fully functional, tests aside. For example, the various `test_xxx.py` scripts in `/examples` do work, provided a local "fix" of the `omfit_classes` as I suggested [here](https://github.com/gafusion/OMFIT-source/issues/6966).

All my changes, both in terms of physics and in terms of new/updated input/output of the code, are also already **extensively documented** in the **docs** of this branch. Please read the new docs before starting the review. Just to summarize it as much as possible:
- The **radial particle flux** at the outermost modelled flux surface interacts with the **main wall reservoir**. This already happened in the current version of Aurora, but before absorbed particles could only be permanently retained, or released with a given time constant. Now, the interaction is much more complex/realisitc. Here, particles can either be **reflected as fast ions**, or be **promptly recycled as thermal ions**, or can be **implanted** in the main wall reservoir (which has a limited capability, which emulates the wall saturation effect, which has to be set by the user). Particles implanted in the main wall can be released through **physical sputtering**. This sputtering can be caused by the radial flux of the simulated impurity itself, and by the flux(es) of the background species. While the first one is self-consistently simulated, the second one(s) must be in some way set by the users. Particles released by the wall (either reflected, thermally released or sputtered) constitute a source for the plasma. The radial profile of this source (i.e. how much they penetrate the plasma before being ionized) is calculated according to the energy at which these are released. Reflection coefficients and reflection energies are avaialble for a numerous amount of impurities (D, He, N, Ne, Ar, ...) and wall materials (Be, C, W), extracted from a citable database. Sputtering yields and sputtering energies for impurities implanted within the wall are now only available for the He-W combination, but we plan to calculate them for other relevant combinations in the future.
- Regarding the **parallel particle flux** (resulting from the parallel loss term in the SOL): in the current version of Aurora, this flux would directly go from the **divertor reservoir**, from which could flow back towards the plasma or being pumped. Now, it will interact with a divertor wall reservoir. This interaction is by all means identical to the interaction of particles with the main wall. The main difference is that the released particles from this reservoir will go towards the divertor reservoir, and only from this moment they can either flow back to the plasma or being pumped.

There are also further additional options regarding the recycling/pumping model:
- It is possible to define a **screening coefficient**, determining how much of the "backflow" from the divertor reservoir really goes back to the plasma, becoming a source for it, alternatively to go back immediately to the divertor wall. This is user-defined.
- It is possible to define a **recombination coefficient**, determining how much of the parallel particle flux really goes towards the divertor wall, undergoing the plasma-wall interaction, or alternatively to go directly towards the divertor reservoir. This is also user-defined.
- It is optionally possible to add a second neutral reservoir, the **pump reservoir**, from which pumping is performed (this was particularly useful for simulating, e.g., the AUG vessel geometry).

This was an extreme summary of the changes included in this PR. I attach here again my usual sketch, which the community has learnt quite well in the last year :) You will find many more sketches, as well as practical examples of all these functionalities, in the docs.

aurora_model

Of course (and I really cannot stress it enough), all these changes are **pure optional additions** to the code, and are **not activated by default**. If a user doesn't want to use this advanced plasma-wall interaction model and this extended recycling/pumping model (or doesn't even know of their existence), the code behavior would be **completely unchanged** wrt the current Aurora master version. All the thousand additional lines of core are completely invisible, if one doesn't explicitly call them. You can test this against your own Aurora simulations, if you want. Also, although the new/modified fluxes and interactions are deep coded in the main fortran algorithms (which are always called, also in the case of "basic" simulations), the presence of these new lines in the fortran files does not affect at all the velocity at which simulations without the new models are executed.

There is **one single backwards incompatibility**, not related to the physics model but with the output of the code. Namely, the main output, i.e. the output of `asim.run_aurora`, **is not anymore a tuple but a dictionary**. We concluded that this was a necessary change, because a tuple was much un-flexible in terms of future devopments (and, like in this case, addition of new output fields). So, for example, now the plasma density after a simulation would be called as `out['nz']` instead of `out[0]`. All the test/example scripts in this PR are already amended accordingly, and the same needs to be done by the user scripts, if/when they will want to update to the new Aurora version.

Here follows a brief summary of the main changes to the various files:
- `core.py`: The various additional entries in the input namelist are loaded, and it is automatically checked if the selected values are allowed. While calling the main Fortran algorithm, among the inputs there are also the user-defined initial values for the density in all the various reservoirs (before it was possible, if I remember correctly, to set an initial value only for the plasma, while all the other ones did start only from 0... this does not really make sense, if one wants to solve an initial-value problem). Additionaly, the fortran algorithm accepts now many more inputs. The output of the call to the main Fortran algorithm is now a dictionary, and not a tuple anymore, and there are many more additional keys related to the new fluxes/reservoirs. The function `check_conservation` has been renamed to `reservoirs_time_traces` (which is more consistent with its content), and additionally to check che particle conservation, it always displays all time traces, including the reservoirs which are not used (in which case they are just always zero). It is important to note that this is default class which is called for simulations without wall model, as before (because, in case the wall model is desired, one has to call `pwi.py`, see later). Therefore, in this class all inputs related to the wall model are set to zero (or consistent with deactivated wall model), and as such are passed to the fortran algorithm.
- `pwi.py`: This is a new file, which is the class which needs to be called (instead of `core.py`) when one wants to use the wall model. It inherits all the methods from `core.py`, and adds new one needed to set up the wall model. In particular, we have the functions:
`setup_PWI_model`: starting from the inputs present in the namelist, it prepares all the fields to pass as input to the fortran algorithm, related to surface coefficients, impact energies, etc., mainly calling functions from the new preparation file `surface.py` (see later).
`get_background_fluxes`: it prepares the fluxes of the background species onto the wall, starting from user inputs, to be passed to the fortran algorithm.
`get_impact_energy`: it calculates the impact energies of the various fluxes onto the walls.
`setup_rec_profs`: it prepares the radial profiles of the sources, in the plasma, coming from neutrals released from the main wall (reflected, thermally released, or sputtered), as function of the energies at which they are released.
And then there is again the call to the main fortran algorithm, which is the same as done in `core.py`, but with inputs related to the wall model not set to zero.
Also, there is the function `PWI_time_traces` which, in the same way as `reservoir_time_traces` does, automatically plots all the main time traces (fluxes, surface densities, etc.) related to the wall model.
- `main.f90`: Here, most of the additional/edited lines are caused bu the numerous additional inputs to, and outputs from, the fortran algorithm. Additionally to that, there is the passing of these new additional inputs to both algorithms `impden0` and `impden1`, and, in the subroutine `edge_model`, we find the additional calculations related to the extension of the recycling/pumping model, i.e., as mentioned before:
The fact that the parallel losses can go towards a divertor wall rather than directly into the divertor reservoir.
The possibility to define a recombination efficiency in the divertor.
The possibility to define a screening efficiency for the backflow from divertor towards plasma.
The possibility to define an additional pump chamber.
The possibility to set physical dimensions to the neutral reservoir(s), which allows to define the pumping through an engineering pumping speed rather than with a pumping time constant.
- `impden.f90`. Only here we finally find the actual calculations which make the wall model. The main idea is that, if the wall model is not activated, the fluxes from/into the walls and the wall surface densities are calculated in the old way. If the wall model is activated, instead, for both algorithms `impden0` and `impden1` the fluxes from/into the walls and the wall surface densities are calculated calling the same subroutine `advanced_PWI_model`, which is put at the end of the file and is made by relatively few lines of code.
- `surface.py`: This is a new file, which contains several functions aimed to read, interpolate and process surface data (reflection coefficients/energies, sputtering yields/energies, etc.), starting from the appropriate datafiles, in order to use them as inputs for the code. Generally speaking, I wrote this inspiring to the atomic data counterpart `atomic.py`.
- `trim_files.py`: This is a new files, which is used to load the appropriate surface datafile for any type of interaction and for any impurity/wall material combination. Also, I wrote this inspiring to the atomic data counterpart `adas_files.py`. The various datafiles are present in the folder `/aurora/trim_files`, and are the same ones used by B2.5, within SOLPS-ITER, for modelling the plasma-surface interactions, and are extracted from an IPP report by Eckstein from 2002 (which is of course adequately cited). This is true for reflection, bulk sputtering and implantation depth. The datafiles for the sputtering yields/energies of impurities implanted within a given wall material, instead, were written on purpose for my PhD work, and, as already said, for now only the combination He-W is present.

All the new possibilities related to the wall model and to the extended recycling/pumping model are also adequately documented in further examples scripts, put in the `/examples` folder. There are also some example scripts for how to read and plot surface coefficients.

I guess that this is it, for now. I also quickly mention what I already have in the to-do list, once the review will be completed:
- Re-introduce, in the online branch, the methods and functions which allow to periodically modulate the transport coefficients to set for the simulation, in order to emulate ELMs, and the various plotting/output functions related to this. Although I used this for my personal investigation, you asked me to leave it outside this PR for now, also in order to alleviate a bit the review job. However, since it was, apparently, the part of my results which caught most the attention of the audience so far, and since I really see this as a fundamental part for future integrated modelling investigation, I really think that it will be worthy to put this online as well.
- Generalize the wall model in order to simulate not only impurities implanted into the wall, but wall impurities as well. Namely, to be able to model the transport, in the plasma, of eroded wall species (C, Be, W, ...) in a self-consistent way, i.e. without the need of specifying an arbitrary edge source to match the experimental data.

I hope you will find this long and detailed summary useful for the review job! :)

PR #81, opened 2023-04-07
Replace numerical integral by analytical in bremsstrahlung  formula

odstrcilt:
Original numerical integral was very slow, and it took several minutes to evaluate for larger Aurora simulations. SOmetimes it even crashed due to large memory requirements.  
Replacing it with a trivial analytical solution made it several orders of magnitude faster.


fsciortino (approved):
This seems like a good change for most uses. Thanks @odstrcilt !

PR #83, opened 2023-05-31
Time dependent NBI CX

odstrcilt:
add support for time-dependent NBI CX rate. It should be backward compatible with previous time-independent input.


fsciortino:
@odstrcilt some tests are failing. Can you please check?

odstrcilt:
It seems to be unrelated to my change, I have no idea how to fix it. Perhaps I could make an empty push to rerun the regression tests again.

odstrcilt:
Based on the error, it looks like the python3.9 installation is missing numpy package.

PR #89, opened 2023-08-21
minor bug in plotting example in facit.py

DanielFajardoJ:
For the plots at the end, the convection variables were being called fct.Vz_ instead of fct.Vconv_, which gave rise to an error of course, so I corrected this.


fsciortino (approved):

fsciortino:
@DanielFajardoJ did you notice that this was failing in the build tests? It should not have been merged. I don't think that it can be "undone". Can you please try to look into why it's failing the build tests? This won't be easier, but I think that it's a good learning experience anyway.

DanielFajardoJ:
@fsciortino I hadn't noticed, sorry. I will have a look. But it is surprising, because I made a new installation of Aurora (for the new IPP file systems) and then only changed two lines in facit.py, related to the example plot at the end, and didn't touch anything else. Is this causing trouble somewhere else?

PR #93, opened 2023-12-29
Generalize FACIT inputs

odstrcilt:
Small change that allowes to running FACIT for nD input arrays, for example, in case of time-dependent transport coefficients

### DIFF
--- a/aurora/coords.py
+++ b/aurora/coords.py
@@ -172,7 +172,6 @@ def rV_vol_average(quant, r_V):
 
     return quant_vol_avg
 
-
 def rad_coord_transform(x, name_in, name_out, geqdsk):
     """Transform from one radial coordinate to another. Note that this coordinate conversion is only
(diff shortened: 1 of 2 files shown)


DanielFajardoJ (commented):
@odstrcilt thank you for the proposed improvements!

I only notice two small details, one related to your changes and one I just saw was there from before.

In line 312:
`JV = np.atleast_2d(JV)`
The Jacobian has not been defined at this point yet. If the code goes into the `if` statement of line 310 then it will go into the statement of line 330, where JV will be defined. So I think we can delete the current line 312.

In lines 1455 to 1458 (plotting the example), all `fct.Vz_{CL,BP,PS}` should be `fct.Vconv_{CL,BP,PS}`. Perhaps we can fix this during this commit?

odstrcilt:
Thank you for checking it. I have fixed both issues. Can I merge it?

DanielFajardoJ (approved):
@odstrcilt looks good, thanks!

PR #94, opened 2024-02-28
caching for adas files loading

fabGitHubTester:
Hi,
This should speed up the loading of the adas files especially in a scenario where we repeat the loading of the same files many times (ex: optimization or uncertainty quantification).

### DIFF
--- a/aurora/atomic.py
+++ b/aurora/atomic.py
@@ -329,7 +329,6 @@ def read_filter_response(filepath, adas_format=True, plot=False):
 
     return E_eV, response
 
-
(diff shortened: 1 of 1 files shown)


fsciortino:
@odstrcilt could you please check whether this may have nasty interactions with the OMFIT environment? If not, I think that we can merge. I wonder if this trick by @fabGitHubTester could have wider applicability in Aurora -- avoiding explicit cashing of ADAS files.

odstrcilt (approved):

PR #96, opened 2024-03-12
Fix CI: teach setuptools that the build needs numpy

eguiraud:
Recent versions of setuptools need the build requirements
to be specified in the pyproject.toml, see
https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#build-system-requirement

Editable installs are still not working, so this patch temporarily
disables them in order to bring the CI back to green.
The issue with editable installs is tracked at #98.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         build_command: ["python setup.py develop",
                         "pip install .",
-                        "pip install --editable .",
+                        #"pip install --editable .", # FIXME: see issue #98
                         "python setup.py install",
                         # "echo 'no operation'",  # how to test this one?
         ]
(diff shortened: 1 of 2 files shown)


fsciortino (approved):
Let's see if this works!

eguiraud-pf:
Ok so this fixes installation of the package except for `--editable` installations. I'll see what I can do about that :)

eguiraud:
@fsciortino see new PR description: I propose to disable editable installs for now so you get a green CI back (although one that does not test editable installs, for now).

I'll propose a larger refactoring of the building of the fortran extension modules soon that should bring editable installs back.

eguiraud:
...never mind, setting up the new build system took less than I expected, I think we can skip this step and directly discuss #99 .

PR #103, opened 2024-07-19
Add titan files

odstrcilt:


odstrcilt:
Hi @fsciortino could you give me permission to merge the request even without reviews if the regression tests pass?

odstrcilt:
@fsciortino ready to be merged

.github/workflows/python-publish.yml
```diff
@@ -20,8 +20,8 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install setuptools wheel numpy
-    - name: Build the package
-      run: python setup.py sdist
+    #- name: Build the package
+    #  run: python setup.py sdist
```
fsciortino, line 24: @odstrcilt merging something like this needs to be done carefully... what's your reasoning here? Are we giving up on distributing via PyPi until we can manage the omfit_classes dependence?

pyproject.toml
```diff
@@ -12,7 +12,7 @@ classifiers = [
 ]
 dependencies = [
     "matplotlib",
-    "numpy",
+    "numpy<2.0.0",
```
fsciortino, line 15: @odstrcilt can you please explain? Why should this be done across the whole package and for all users?

fsciortino (changes requested):
Some small changes that seem hacks to get the regression tests to change. We can go with this intentionally, but we should then have a clear plan on how to go back to PyPi publishing and rigorous depedencies management.

.github/workflows/python-publish.yml
odstrcilt: Because the setup.py file is no longer there, Enrico has removed it. The regression tests passed, it it has somehow installed, so I expect that it works.

pyproject.toml
odstrcilt: This was the only option how to make the regression test pass. It has produced this error:
https://github.com/fsciortino/Aurora/actions/runs/10012842877/job/27679164044?pr=103
Enrico suggested using the older numpy 
https://github.com/fsciortino/Aurora/pull/99#issuecomment-2240023051

PR #106, opened 2024-08-08
Finish work on new PWI model

odstrcilt:
I have made several changes to the new PWI model. 

- Merge with all changes which have accumulated over time in AURORA
- Remove code duplications. Mainly the run function was removed from  aurora_sim_pwi class
- Simplify the code if possible to keep the code readable and easy for further expansion.  

Changes compared to the original new_recycling_PWI_model branch are here: 
https://github.com/fsciortino/Aurora/compare/new_recycling_PWI_model...new_recycling_PWI_model_merged

I have check the examples and the results looks identical. 
New model is 20% slower than the original one. Any idea why? It should not calculate anything extra?


odstrcilt:
Thank you for excellent work, @antonellozito, and I apologize that it took me so long to finish it. It looks much better than the first version. I have removed most of the duplicated code and simplified the scripts, where possible. 
I'm also thinking about remove all plotting function from core.py and moving them to a separate script. But this can be done some time later.

antonellozito:
Thanks for the review @odstrcilt! In the next days I will double check it, and see if also my personal scripts still do what they did before your changes (although I assume it will be the case).

antonellozito (approved):
@odstrcilt Thanks for the work!! To what I can see, the output is unchanged wrt my original version, even in case of full PWI model activated, so from my side this is ok to go.
A little side request: considering that, after the last commits, the installation method of the package is changed (e.g. moving from a Makefile to a CMakeLists.txt), could you please update the README accordingly? It still makes reference to the old setup.py, hence not explaining properly how to install Aurora in case of Git cloning.

PR #118, opened 2025-03-10
Update python-publish.yml

odstrcilt:
try to fix the pip instalation. Likely caused by obsolete yaml file

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -10,6 +10,8 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write  # 🔥 REQUIRED for Trusted Publishing
     steps:
     - uses: actions/checkout@v2
     - name: Set up Python
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -36,8 +36,8 @@ jobs:
 
     - name: Publish distribution 📦 to Test PyPI
       run: |
-        python --version
-        auditwheel repair dist/aurorafusion*.whl
+        #python --version
+        #auditwheel repair dist/aurorafusion*.whl
         twine upload --repository testpypi wheelhouse/*  --verbose
 
         #twine upload --repository-url https://test.pypi.org/legacy/ dist/* --verbose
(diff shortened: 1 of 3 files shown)


cjperks7:
@odstrcilt Do you have a minimum working example or a suggestion to check that this worked? I vaguely understand that "pip installation"'s live on a dedicated public repo website so are these changes getting pushed there as well so I can check with "pip install aurora ..." or are only specific steps in version # posted? I don't know about this stuff so I want to avoid just blindly accepting pull request for every incremental fix that needs to happen and get this done in one go. 

Can what you're doing here be quickly/conveniently explained? If you know for certain that this is the last thing that needs to happen and everything is good then that's fine too and you can disregard all this extra work

odstrcilt:
The ultimate check is that it will change a version here:
https://pypi.org/project/aurorafusion/

I'm starting to understand how it works. 

publishing is done by 
 https://github.com/fsciortino/Aurora/tree/master/.github/workflows/python-publish.yml


here is the output from this workflow
https://github.com/fsciortino/Aurora/actions/runs/13776328715/job/38526209923

it works, but the publishing was skipped 

publishing happens if this condition is satisfied 
startsWith(github.ref, 'refs/tags')

and this happens if the new tag is created 

https://github.com/fsciortino/Aurora/tags

the last tag was not working because the python-publish.yml was obsolete

I have tried to fix it, hopefully it will work this time.

cjperks7:
That makes sense and thanks for explaining! Seems like you know what you're doing with this "Upload Python Package / deploy (push)" check that's running. Maybe you've already seen this, but looks like since your last commit "267fcad" it caused that check to fail. I reran it this morning to make sure it wasn't just some weird one-off thing since your commit "8635ac1" passed.

From what I see in the output, looks like there's a small issues with line 31 "repository_url: https://test.pypi.org/legacy/" in "python-publish.yml"

odstrcilt:
@cjperks7 I fixed it!!! it took me the whole day, but I have found a working solution which has passed the PyPi upload test. 
https://github.com/fsciortino/Aurora/actions/runs/13798628317/job/38596123374

I will add back the condition, that the publishing happens only if the git tag changes. Could you merge it? I hope that the last time.

cjperks7 (approved):
Wow, this looked like a special kind of hell. Good job!

PR #120, opened 2025-03-12
Another try to fix publishing

odstrcilt:
move one step back, use twine to publish the package, it was working last time.


odstrcilt:
Thnak you;

cjperks7 (approved):
I don't really know how python publishing works. Seemed like all checks successfully passed last PR, but an error popped up after the merge. I'm just going to go ahead and merge this, but if the same problem as last time happens maybe provide a minimum working example to better facilitate iteration

